PR #110, opened 2016-11-04
Table formatter and command specific formatters

### LINKED ISSUE #105: Implement table output formatting.
There should be a formatter to transform maps returned from `list_*` commands to the table format like in the old CLI.

hairyhum:
Fixes #105
Implemented table formatter to output `list_*` data the same way that old CLI does.

Refactored command run code to allow commands to specify default formatter (like in list_queues_command)

Removed error handling from formatters and printers. Errors are now handled in `rabbitmqctl` module.

Errors are printed to STDERR.


michaelklishin:
After merging #111 via master into this branch I only have the 4 failures above.

PR #116, opened 2016-11-15
CSV formatter

### LINKED ISSUE #115: Implement CSV formatter
To export `list_*` commands output to file it would more convenient to use CSV standard, instead of tab-separated table formatter.

Formatter should support output from `list_*` commands and return info items names in a first row.

hairyhum:
Fixes #115
Flattens `list_consumers`, outputs map keys as a first row. Info items are formatted the same way as with table formatter.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,10 +1,11 @@
 PROJECT = rabbitmq_cli
 VERSION ?= 0.0.1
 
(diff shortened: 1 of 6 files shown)


michaelklishin:
I ❤️ this feature.

PR #122, opened 2016-11-18
Add encode command

### LINKED ISSUE #118: Add encode command
See https://github.com/rabbitmq/rabbitmq-server/issues/979

acogoluegnes:
Catching up with rabbitmq/rabbitmq-server#979.

Fixes #118

Should be merged once https://github.com/rabbitmq/rabbitmq-server/pull/1032 and https://github.com/rabbitmq/rabbitmq-common/pull/146 are merged into both stable and master.

### DIFF
--- a/lib/rabbitmq/cli/ctl/commands/encode_command.ex
+++ b/lib/rabbitmq/cli/ctl/commands/encode_command.ex
@@ -62,9 +62,6 @@ defmodule RabbitMQ.CLI.Ctl.Commands.EncodeCommand do
         decode: :boolean
       ]
     end
-  def aliases(), do: []
-
-  def scopes(), do: [:ctl]
 
   def run(_, %{list_ciphers: true}) do
     {:ok, :rabbit_pbe.supported_ciphers()}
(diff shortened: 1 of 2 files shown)


lib/rabbitmq/cli/ctl/commands/encode_command.ex
```diff
@@ -0,0 +1,135 @@
...
+        list_hashes: :boolean,
+        cipher: :atom,
+        hash: :atom,
+        iterations: :integer,
+        decode: :boolean
+      ]
+    end
+  def aliases(), do: []
+
+  def scopes(), do: [:ctl]
```
hairyhum: `scopes/0` callback is optional. Default scope is calculated from a module name. For `RabbitMQ.CLI.Ctl` namespace it will be `:ctl`.

lib/rabbitmq/cli/ctl/commands/encode_command.ex
```diff
@@ -0,0 +1,135 @@
...
+    end
+  end
+
+  def formatter(), do: RabbitMQ.CLI.Formatters.Erlang
+
+  def usage, do: "[--decode] [value] [passphrase] [--list-ciphers] [--list-hashes] [--cipher cipher] [--hash hash] [--iterations iterations]"
+
+  def flags, do: Keyword.keys(switches())
+
+  def banner(_, _), do: ""
```
hairyhum: There are several uses of this command, so it's better to print what it's going to do.

lib/rabbitmq/cli/ctl/commands/encode_command.ex
```diff
@@ -0,0 +1,135 @@
...
+
+  def evaluate_input_as_term(input) do
```
hairyhum: Should this function be exported?

michaelklishin: If it's not used by tests, I agree.

lib/rabbitmq/cli/ctl/commands/encode_command.ex
```diff
@@ -0,0 +1,135 @@
...
+      [
+        list_ciphers: :boolean,
+        list_hashes: :boolean,
+        cipher: :atom,
+        hash: :atom,
+        iterations: :integer,
+        decode: :boolean
+      ]
+    end
+  def aliases(), do: []
```
michaelklishin: Note that they can but now only after https://github.com/rabbitmq/rabbitmq-cli/pull/128 is merged.

lib/rabbitmq/cli/ctl/commands/encode_command.ex
michaelklishin: I concur.

PR #124, opened 2016-11-21
Make switches as aliases callbacks optional.

### LINKED ISSUE #120: Make `switches/0` an optional callback
After https://github.com/rabbitmq/rabbitmq-cli/commit/59ba4d91b379fb76e829f7d037f6c9a04146e34f it becomes confusing that some commands won't have to define any switches since they only support `-p [vhost]` (which is auto-defined by the tool).

[delete-shovel](https://github.com/rabbitmq/rabbitmq-shovel/blob/master/src/Elixir.RabbitMQ.CLI.Ctl.Commands.DeleteShovelCommand.erl#L41) is a good example of this.

The `switches/0` callback therefore should probably be made optional.

hairyhum:
Fixes #120 

Refactored parser to parse command specific and global flags separately, so there is no need to load commands before parsing.
Should also make the CTL faster

### DIFF
--- a/lib/rabbitmq/cli/auto_complete.ex
+++ b/lib/rabbitmq/cli/auto_complete.ex
@@ -24,7 +24,7 @@ defmodule Rabbitmq.CLI.AutoComplete do
     case List.last(tokens) do
       nil        -> [];
       last_token ->
-        {args, opts, _} = Parser.parse(tokens)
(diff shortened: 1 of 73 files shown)


michaelklishin:
Reverted because of https://github.com/rabbitmq/rabbitmq-cli/issues/126.

Issue #126, opened 2016-11-22
Global options parsing cannot be relied on to parse positional arguments correctly


michaelklishin:
If you trace how `set_policy --priority 1` is parsed it turns out to be

```
./scripts/rabbitmqctl set_policy --priority 1 federate-me "^amq." '{"federation-upstream-set":"all"}'
** (CaseClauseError) no case clause matching: {["set_policy", "federate-me", "^amq.", "{\"federation-upstream-set\":\"all\"}"], %{priority: 1}, []}
    (rabbitmqctl) lib/rabbitmqctl.ex:43: RabbitMQCtl.main/1
    (elixir) lib/kernel/cli.ex:76: anonymous fn/3 in Kernel.CLI.exec_fun/2
```

```
{"set_policy", RabbitMQ.CLI.Ctl.Commands.SetPolicyCommand, ["1", "federate-me", "^amq.", "{\"federation-upstream-set\":\"all\"}"]}
```

so the "1" argument gets treated as positional. This is wrong and breaks existing commands.

`Elixir.OptionParser`, however, is not to blame:

```

iex(4)> OptionParser.parse(["--priority=1"], switches: [priority: :integer])
{[priority: 1], [], []}
iex(5)> OptionParser.parse(["--priority", "1"], switches: [priority: :integer])
{[priority: 1], [], []}
```

michaelklishin:
I believe this started with 9196ef719a7261d5c7e86f4087f3e9fb41e97520.

michaelklishin:
There turns out to be a fundamental problem with https://github.com/rabbitmq/rabbitmq-cli/commit/9196ef719a7261d5c7e86f4087f3e9fb41e97520: "global" options parser cannot be relied on to parse positional arguments because it is unaware of command-specific switches and will leave extra values behind.

The only fix I could come up with in a couple of hours includes ignoring positional arguments from the global parser. This also has a downside: with that we effectively require command name to be the first argument but [rabbitmqctl(1)](https://www.rabbitmq.com/man/rabbitmqctl.1.man.html) suggests and encourages global options to be provided first :(

michaelklishin:
I'm going to revert https://github.com/rabbitmq/rabbitmq-cli/pull/124, which fixes this.

PR #128, opened 2016-11-22
Make switches/0 and aliases/0 optional command callbacks

### LINKED ISSUE #120: Make `switches/0` an optional callback

michaelklishin:
Fixes #120. Same as #124 but doesn't include any unrelated changes
and thus suffer from #126.

### DIFF
--- a/lib/rabbitmq/cli/command_behaviour.ex
+++ b/lib/rabbitmq/cli/command_behaviour.ex
@@ -20,8 +20,6 @@ defmodule RabbitMQ.CLI.CommandBehaviour do
   @callback merge_defaults(List.t, Map.t) :: {List.t, Map.t}
   @callback banner(List.t, Map.t) :: String.t
   @callback run(List.t, Map.t) :: any
-  @callback switches() :: Keyword.t
-  @callback aliases() :: Keyword.t
   # Coerces run/2 return value into the standard command output form
   # that is then formatted, printed and returned as an exit code.
   # There is a default implementation for this callback in DefaultOutput module
@@ -29,7 +27,13 @@ defmodule RabbitMQ.CLI.CommandBehaviour do
                                   {:error, ExitCodes.exit_code, [String.t]}
   @optional_callbacks formatter: 0,
                       scopes: 0,
-                      usage_additional: 0
+                      usage_additional: 0,
+                      switches: 0,
+                      aliases: 0
+
+  @callback switches() :: Keyword.t
+  @callback aliases() :: Keyword.t
+
   @callback formatter() :: Atom.t
   @callback scopes() :: [Atom.t]
   @callback usage_additional() :: String.t | [String.t]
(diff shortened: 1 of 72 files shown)


lib/rabbitmq/cli/command_behaviour.ex
```diff
@@ -20,16 +20,20 @@ defmodule RabbitMQ.CLI.CommandBehaviour do
...
-  @callback aliases() :: Keyword.t
   # Coerces run/2 return value into the standard command output form
   # that is then formatted, printed and returned as an exit code.
   # There is a default implementation for this callback in DefaultOutput module
   @callback output(any, Map.t) :: :ok | {:ok, any} | {:stream, Enum.t} |
                                   {:error, ExitCodes.exit_code, [String.t]}
   @optional_callbacks formatter: 0,
                       scopes: 0,
-                      usage_additional: 0
+                      usage_additional: 0,
```
acogoluegnes: Is there a way to provide a default implementation for an optional callback? The [documentation](http://elixir-lang.org/docs/stable/elixir/behaviours) mentions it, but I couldn't find a proper example and couldn't come up with something that would work.

michaelklishin: There is no inheritance in Elixir and this is not a problem with polymorphism, so protocols do not fit either.

michaelklishin: [This blog post about Elixir 1.3](http://tuvistavie.com/2016/elixir-1-3/#support-of-otp-optionalcallback) mentions default implementations but I cannot find any specific examples.

michaelklishin: I guess the point of "default implementation" means that a behaviour could also define a function to fall back to but that wouldn't help us avoid `function_exported?/4` entirely.

acogoluegnes: OK. I let @hairyhum have a look at the PR before merging.

hairyhum: There is something like "default implementation" by using [Elixir mixins](http://elixir-lang.org/getting-started/alias-require-and-import.html#use)
But that would work for Elixir commands only, and Erlang commands will have to define all the callbacks.

michaelklishin:
Anything specific left to address in this PR?

hairyhum:
That should be all

michaelklishin:
Thanks @hairyhum.

PR #132, opened 2016-11-30
Support multiple plugins directories

### LINKED ISSUE #131: Support multiple plugins directories.
In order to merge https://github.com/rabbitmq/rabbitmq-server/pull/1016 into master, the CLI should support multiple plugins directories in `RABBITMQ_PLUGINS_DIR`.

acogoluegnes:
Fixes #131


lib/rabbitmq/cli/core/helpers.ex
```diff
@@ -118,23 +119,30 @@ defmodule RabbitMQ.CLI.Core.Helpers do
...
     end
   end
 
+  defp try_load_plugins_from_directory(directory_with_plugins_inside_it) do
+    with {:ok, files} <- File.ls(directory_with_plugins_inside_it)
+    do
+        Enum.filter_map(files,
+            fn(filename) -> String.ends_with?(filename, [".ez"]) end,
+            fn(archive) ->
+              case Regex.named_captures(~r/(?<name>.+)-(?<version>.+).ez/, archive) do
```
binarin: Code in rabbitmq-server is unpacking .ez and fetching version from .app-file within. Is just parsing filename sufficient in this case?

lib/rabbitmq/cli/plugins/commands/list_command.ex
```diff
@@ -114,6 +114,23 @@ defmodule RabbitMQ.CLI.Plugins.Commands.ListCommand do
       plugins: format_plugins(plugins, format, enabled, enabled_implicitly, running)}
   end
 
+  defp remote_running_plugins(node) do
```
binarin: `rabbitmq-server` implementation was split into 2 rpc calls only for compatibility. Maybe it's better to add new function to `rabbit_plugins` that'll return a list of running plugins already augment with versions - as there is nothing yet to preserve backwards compatibility with.

lib/rabbitmq/cli/core/helpers.ex
hairyhum: The next line ignores version, but it still makes sense to get the application name from `.app` file.

acogoluegnes: The code above is used to load plugins code in the command line Elixir app (note it doesn't use the `version` variable). The CLI does a RPC call to `rabbit_plugins:list/1`, which indeed unpacks the EZ file to fetch the version from the `src` file.

PR #133, opened 2016-11-30
Load command specific options from a single command

### LINKED ISSUE #129: Support overlapping switches in commands
Currently command specific switches are merged for all commands and processed in `parse`.

So commands with overlapping switch names (or aliases) will have execution errors.

hairyhum:
Fixes #129 
Change the command parsing behaviour to return a command if it was found, or `:no_command` if it wasn't.
Move responsibility for command discovery from `rabbitmqctl` to `parser`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -56,7 +56,7 @@ test:: all
 
 clean::
 	rm -f $(ESCRIPTS)
-	- mix local.hex --force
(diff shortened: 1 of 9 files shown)


michaelklishin:
So this is another take on all the "extras" in #124?

hairyhum:
@michaelklishin yes. Now moving everything to `parser.ex` to avoid confusion.

PR #137, opened 2016-12-05
Fix longname-mode on hosts without detectable FQDN

binarin:
Re-implement https://github.com/rabbitmq/rabbitmq-server/issues/890

CLI tools fail in longnames-mode if erlang is not able to determine host
FQDN (with at least one dot in it). E.g. this happens on docker.

And it was not possible to alleviate this situation using any options
from http://erlang.org/doc/apps/erts/inet_cfg.html


michaelklishin:
Thank you!

PR #138, opened 2016-12-05
Handle errors from distribution start attempts

binarin:
Otherwise broken or misconfigured distribution is reported as an
unavailable broker node.

### DIFF
--- a/lib/rabbitmqctl.ex
+++ b/lib/rabbitmqctl.ex
@@ -25,6 +25,9 @@ defmodule RabbitMQCtl do
   import RabbitMQ.CLI.Core.Helpers
   import  RabbitMQ.CLI.Core.Parser
 
+  @type options() :: Map.t
+  @type command_result() :: {:error, ExitCodes.exit_code, term()} | term()
+
   def main(["--auto-complete", "./rabbitmqctl " <> str]) do
     auto_complete(str)
   end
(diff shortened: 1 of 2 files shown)


lib/rabbitmqctl.ex
```diff
@@ -229,4 +233,18 @@ defmodule RabbitMQCtl do
...
     exit({:shutdown, code})
   end
+
+  @spec with_distribution(options(), (() -> command_result())) :: command_result()
+  defp with_distribution(options, code) do
+    # Tries to start net_kernel distribution, and calls `code`
+    # function on success. Otherswise returns error suitable for
+    # handle_shutdown/0.
+    case Distribution.start(options) do
+      :ok ->
```
michaelklishin: `Distribution.start/1` returns a tuple of 2 (because `net_kernel:start/1` does).

binarin: This is unpacked at 
https://github.com/binarin/rabbitmq-cli/blob/71d753b8088317d91763c0103fcec4d5b7027ce2/lib/rabbitmq/cli/core/distribution.ex#L49

michaelklishin: Well, with this PR as is I cannot start a test node for this very repository. I think I have a couple of changes that get it to work, though.

michaelklishin: I've pushed a couple of corrections that pass all tests. Check them out, maybe it's some of the things you have in mind.

PR #139, opened 2016-12-06
Real time output in enable/disable/set commands

### LINKED ISSUE #134: Make rabbitmq-plugins print info in realtime (as a stream)
rabbitmq-plugins tool prints all info after `run` is finished. As an operator I want to see that plugins are going to be started and are being started when it happens in real time (like in the old cli)

dcorbacho:
Closes #134


hairyhum:
There should be some merge/rebase issue, because I see `augment_version` function removed in the diff.
@dcorbacho can you check the changes you've made and apply them to the latest `master`?

dcorbacho:
@hairyhum it should be solved now

hairyhum:
Tests need to be updated. 17 tests expect maps, but get streams.

PR #156, opened 2016-12-13
Load .ez plugins without version

### LINKED ISSUE #153: Plugin discovery: Support plugins without a version.
Plugin archive name is matched with `/(?<name>.+)-(?<version>.+).ez/`, which assumes a version for each plugin and then uses this version for a directory name inside the archive.

The CLI should support other formats and load version from an application file.

dcorbacho:
Closes #153


acogoluegnes:
Creating a couple of other test plugins would be a good idea. New tests shouldn't use the existing plugins though: making several tests depending on the same fixture coupled the tests together, one change to a test can affect all the others.

lib/rabbitmq/cli/core/helpers.ex
```diff
@@ -127,17 +129,36 @@ defmodule RabbitMQ.CLI.Core.Helpers do
...
+                ebin_dir |> Code.append_path()
+                app_name |> String.to_atom() |> Application.load()
+            end
+          end)
+    end
+  end
+
+  defp find_dot_app([head | tail]) when Record.is_record(head, :zip_file) do
+    name = :erlang.element(2, head)
+    case Regex.match?(~r/(.+)\/ebin\/(.+)\.app$/, to_string name) do
```
gerhard: Is `(.+)\/ebin\/` required? All tests pass if I change the Regexp to:

```elixir
Regex.match?(~r/(.+)\.app$/, to_string name)
```

gerhard: 👍

PR #163, opened 2017-01-06
Close all connections command

### LINKED ISSUE #155: Introduce a command to close all or a subset of connections
Similar to the idea in https://github.com/rabbitmq/rabbitmq-server/issues/817 but to force close all active connections, in a particular vhost or (worth considering) globally. This is sometimes useful for technical ops.

This will be a good feature to introduce now that https://github.com/rabbitmq/rabbitmq-server/issues/500 is finished and merged.

More ideas suggested by @johnfoldager in the [original issue in rabbitmq-server](https://github.com/rabbitmq/rabbitmq-server/issues/849):

 * a per-connection delay to avoid/lessen the probability of causing a thundering herd, e.g. `--per-connection-delay` in milliseconds
 * a `--global` flag that would close *all active connections* to a node (across all vhosts)
 * a `--limit` of how many connections are closed (cannot work with `--global`, only per-vhost)

dcorbacho:
Closes #155 with https://github.com/rabbitmq/rabbitmq-server/pull/1074

@michaelklishin  I'm not sure that the semantics of how to pair valid options are clear (vhost & limit, node & global, delay with any), any suggestions?

```close_all_connections [-p <vhost> --limit <limit>] [-n <node> --global] [--per-connection-delay <delay>] <explanation>```

### DIFF
--- a/lib/rabbitmq/cli/ctl/commands/close_all_connections_command.ex
+++ b/lib/rabbitmq/cli/ctl/commands/close_all_connections_command.ex
@@ -63,12 +63,12 @@ defmodule RabbitMQ.CLI.Ctl.Commands.CloseAllConnectionsCommand do
   def usage, do: "close_all_connections [-p <vhost> --limit <limit>] [-n <node> --global] [--per-connection-delay <delay>] <explanation>"
 
   def banner([explanation], %{node: node_name, global: true}) do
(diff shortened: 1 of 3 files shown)


michaelklishin:
@dcorbacho I think what this PR has is reasonable. I doubt complex combinations of switches will be widely used. We can introduce more restrictions later based on user feedback.

PR #175, opened 2017-02-09
Makefile: New `install` target to copy escripts

dumbbell:
... to the directory specified with `$(PREFIX)`.

By default, the source directory is copied using pax(1) so hardlinks are preserved. One can set `USE_SYMLINKS_IN_ESCRIPTS_DIR` to anything to use symlinks instead of hardlinks.

**v2**: The `Makefile` now distinguishes real escripts from linked escripts. So instead of using pax(1) in the new `install` target, real escripts are copied and linked escripts are recreated.

[#138925175]

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -22,9 +22,36 @@ include erlang.mk
 ERLANG_MK_REPO = https://github.com/rabbitmq/erlang.mk.git
 ERLANG_MK_COMMIT = rabbitmq-tmp
 
-ESCRIPTS = escript/rabbitmqctl \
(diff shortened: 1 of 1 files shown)


dumbbell:
This new target is used in rabbitmq/rabbitmq-server#1101 and rabbitmq/rabbitmq-server-release#19 to install escripts in the broker directory (the former pull request) to make them available during development and to testsuites, and in packages (the latter pull request).

In the broker, it would probably be nice if at some point, we used escripts in rabbitmq-cli directly instead of copying them.

PR #186, opened 2017-06-02
Wait in stages.

hairyhum:
Make wait command work in stages with timeouts.

- First wait for a pid file to appear (times out)
- Then wait for erlang distribution to start on the node (times out)
- Then wait for application to be running in the node (does not time out)

Default timeout is set to 10 seconds, which should be enough to
write a pid file and start distribution.

### DIFF
--- a/lib/rabbitmq/cli/core/helpers.ex
+++ b/lib/rabbitmq/cli/core/helpers.ex
@@ -205,18 +205,56 @@ defmodule RabbitMQ.CLI.Core.Helpers do
   # Streamify function sequence.
   # Execution can be terminated by an error {:error, _}.
   # The error will be the last element in the stream.
-  def stream_until_error(funs) do
+  # def stream_until_error(funs) do
+  #   Stream.transform(
+  #     funs, :just,
+  #     fn(f, :just) ->
+  #         case f.() do
+  #           {:error, _} = err -> {[err], :nothing};
+  #           other             -> {[other], :just}
+  #         end;
+  #       (_, :nothing) ->
+  #         {:halt, :nothing}
+  #   end)
+  # end
+
+  # Streamify a function sequence passing result
+  # Execution can be terminated by an error {:error, _}.
+  # The error will be the last element in the stream.
+  # Functions can return {:ok, val}, so val will be passed
+  # to then next function, or {:ok, val, output} where
+  # val will be passed and output will be put into the stream
+  def stream_until_error_parametrised(funs, init) do
     Stream.transform(
-      funs, :just,
-      fn
-        (f, :just) ->
-          case f.() do
+      funs, {:just, init},
+      fn(f, {:just, val}) ->
+          case f.(val) do
             {:error, _} = err -> {[err], :nothing};
-            other             -> {[other], :just}
+            :ok               -> {[], {:just, val}};
+            {:ok, new_val}        -> {[], {:just, new_val}};
+            {:ok, new_val, out}   -> {[out], {:just, new_val}}
           end;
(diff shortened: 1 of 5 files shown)


dumbbell:
I executed `rabbitmqctl wait /tmp/pid` with `1` in `/tmp/pid`. A node is running, but I Ctrl+C'd it once so it's paused.

Here is what I get:
```
$ time ./scripts/rabbitmqctl wait /tmp/pid -t 2
Waiting for node rabbit@rosetta ...
Waiting for pid file '/tmp/pid' to appear
pid is 1
Waiting for erlang distribution on node 'rabbit@rosetta' while OS process '1' is running
Error:
{:timeout, 2000}

real	1m41.238s
user	0m0.697s
sys	0m0.310s
```

1. Before running the actual command, it tries to contact the node already, which doesn't take the timeout into account. It takes about 20 seconds before it displays the following line:

    ```
    Waiting for node rabbit@rosetta ...
    ```

    Then another 5 seconds before it gives up and moves on with the actual command.

2. The rest of the time is spent between:

    ```
    Waiting for erlang distribution on node 'rabbit@rosetta' while OS process '1' is running
    ```
    and:
    ```
    Error:
    {:timeout, 2000}
    ```

Is it expected that it tries to contact the node as a very first step before the executing the command?

About that pidfile, honestly, I'm wondering why we check it entirely in this command. We could simply loop on the second part (contacting the node and checking the running applications) and we would achieve the same behavior. It would even be better because it would work with `rabbitmq-server -detached` which doesn't create a pidfile. I'm not even talking about the situation where the pidfile doesn't match the actual VM PID (like in my example).

hairyhum:
The reason why waiting for pidfile to have a running PID and waiting for distribution to start are different stages is that we can know which stage has failed.
We can of course remove the pidfile argument and just call `wait`. But I don't know why did we had it here in the first place, so I'm not sure how breaking is that change.

hairyhum:
As for your test, if you suspend the node it will not respond to `ned_adm:ping` until the kernel heartbeat timeout, I suppose.
`net_adm:ping` is needed here to be able to wait in loop with timeout.
`await_startup` can also fail with an error, and we would have to distinguish between stop and repeat errors. And I would like to avoid that.

hairyhum:
I think the problem with long wait here is that `wait_for` waits for 10 times instead of taking actual function execution time to account. I'll fix that.

dumbbell:
Thanks, your last commit improved the waiting loop: if the pidfile doesn't exist, it stops after the specified timeout, and if the node doesn't respond (because it's suspended) or is not running, it also stops after the specified timeout (a bit more if the node doesn't respond because it waits for the TCP connect timeout, but that's fine).

What I don't understand is the `Waiting for node rabbit@rosetta ...` before we check the pidfile. Is there a reason to do that?

hairyhum:
It's a banner message, I don't think we need that. 
@dumbbell do you think we need to suppress  log messages ("waiting for X") if `--quet` flag is set?

dumbbell:
> It's a banner message, I don't think we need that.

Not only a banner: it must try to contact the because it takes about 25 seconds on that line if the node is suspended.

> do you think we need to suppress log messages ("waiting for X") if  `--quet` flag is set?

Yes, it should be totally silent, good idea.

hairyhum:
Well there is a separate issue, when CLI is trying to connect to the node before running a command https://github.com/rabbitmq/rabbitmq-cli/issues/189

dumbbell:
About the necessity or usefulness of the pidfile check, this is out of scope for this story anyway. But I'd like to eventually get rid of it.

PR #209, opened 2017-07-20
Recompile regexes at runtime.

### LINKED ISSUE #208: FunctionClauseError in OTP 20
When running a pre-compiled version of the clicompiled with a prior version of erlang/OTP (e.g. 19.3) the cli errors with the following:

```
** (FunctionClauseError) no function clause matching in RabbitMQ.CLI.Core.CommandModules.to_snake_case/1
    (rabbitmqctl) lib/rabbitmq/cli/core/command_modules.ex:138: RabbitMQ.CLI.Core.CommandModules.to_snake_case(nil)
    (rabbitmqctl) lib/rabbitmq/cli/core/command_modules.ex:178: RabbitMQ.CLI.Core.CommandModules.command_scopes/1
    (rabbitmqctl) lib/rabbitmq/cli/core/command_modules.ex:163: RabbitMQ.CLI.Core.CommandModules.command_in_scope/2
    (elixir) lib/enum.ex:814: anonymous fn/3 in Enum.filter/2
    (elixir) lib/enum.ex:1755: Enum."-reduce/3-lists^foldl/2-0-"/3
    (elixir) lib/enum.ex:814: Enum.filter/2
    (rabbitmqctl) lib/rabbitmq/cli/core/command_modules.ex:91: RabbitMQ.CLI.Core.CommandModules.make_module_map/2
    (rabbitmqctl) lib/rabbitmq/cli/core/command_modules.ex:40: RabbitMQ.CLI.Core.CommandModules.load/1
```

kjnilsson:
Because the regex engine may change between OTP versions it is necessary
to recompile any compiled regexes at runtime in case the engine is not
compatible.

Fixes #208 

[#149227527]

### DIFF
--- a/lib/rabbitmq/cli/core/command_modules.ex
+++ b/lib/rabbitmq/cli/core/command_modules.ex
@@ -87,9 +87,10 @@ defmodule RabbitMQ.CLI.Core.CommandModules do
   end
 
   defp make_module_map(modules, scope) do
+    commands_ns = Regex.recompile!(@commands_ns)
     modules
     |> Enum.filter(fn(mod) ->
-                     to_string(mod) =~ @commands_ns
+                     to_string(mod) =~ commands_ns
                      and
                      module_exists?(mod)
                      and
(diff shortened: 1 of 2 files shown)


michaelklishin:
Good idea. I will do that.

Issue #207, opened 2017-07-17
Create a command to report "unresponsive" queues.


hairyhum:
`list_queues` command requests information from each queue process and can hang forever (or fail with a timeout) if a queue process does not respond. We can introduce a new command or a flag to report which queues are not responding but alive.

For example:
```
$ rabbitmqctl list_queues name status
queue_1 running
queue_2 blocked
queue_3 not_responding
queue_4 running
...
```

Or as a separate command:
```
$ rabbitmqctl not_responding_queues
queue_4 <some queue data>
```

hairyhum:
Fixed by #212

PR #212, opened 2017-07-27
Command to report unresponsive queues

dcorbacho:
This command will wait for every single queue to answer within `<queue-timeout>` milliseconds. If no answer is given, will query the next one. This could potentially be done in parallel, but we risk spawning thousands of processes in some environments. Shall we still do it? 

Part of #207

### DIFF
--- a/lib/rabbitmq/cli/ctl/commands/list_unresponsive_queues_command.ex
+++ b/lib/rabbitmq/cli/ctl/commands/list_unresponsive_queues_command.ex
@@ -51,7 +51,12 @@ defmodule RabbitMQ.CLI.Ctl.Commands.ListUnresponsiveQueuesCommand do
   def switches(), do: [queue_timeout: :integer, local: :boolean]
 
   def usage() do
-    "list_unresponsive_queues [--local] [--queue-timeout <queue-timeout>] [<queueinfoitem> ...]"
+    "list_unresponsive_queues [--local] [--queue-timeout <queue-timeout>] [<unresponsiveq_ueueinfoitem> ...]"
+  end
+
+  def usage_additional() do
+      "<unresponsive_queueinfoitem> must be a member of the list [" <>
+      Enum.join(@info_keys, ", ") <> "]."
   end
 
   def run(args, %{node: node_name, vhost: vhost, timeout: timeout,
--- a/lib/rabbitmq/cli/ctl/commands/list_unresponsive_queues_command.ex
+++ b/lib/rabbitmq/cli/ctl/commands/list_unresponsive_queues_command.ex
@@ -15,8 +15,10 @@
 
 
 defmodule RabbitMQ.CLI.Ctl.Commands.ListUnresponsiveQueuesCommand do
+  require RabbitMQ.CLI.Ctl.InfoKeys
   require RabbitMQ.CLI.Ctl.RpcStream
 
+  alias RabbitMQ.CLI.Ctl.InfoKeys, as: InfoKeys
   alias RabbitMQ.CLI.Ctl.RpcStream, as: RpcStream
   alias RabbitMQ.CLI.Core.Helpers, as: Helpers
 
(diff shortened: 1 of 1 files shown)


lib/rabbitmq/cli/ctl/commands/list_unresponsive_queues_command.ex
```diff
@@ -0,0 +1,60 @@
...
+      {chunks, mfas} = case local_opt do
+          true  -> {1, [local_mfa]};
+          false -> {Kernel.length(nodes), [all_mfa]}
+        end
+        RpcStream.receive_list_items(node_name, mfas, timeout, [:name], chunks)
+      end)
+  end
+
+  defp default_opts() do
+    %{vhost: "/", local: false, queue_timeout: 30000}
```
hairyhum: Isn't 30 seconds timeout too big?

michaelklishin: Is this per queue or total?

michaelklishin: Looks like it's per queue. We can lower it to `15s` but going lower can produce false positives.

lib/rabbitmq/cli/ctl/commands/list_unresponsive_queues_command.ex
```diff
@@ -0,0 +1,60 @@
...
+  def run([], %{node: node_name, vhost: vhost, timeout: timeout,
+                queue_timeout: queue_timeout, local: local_opt}) do
+    Helpers.with_nodes_in_cluster(node_name, fn(nodes) ->
+      local_mfa  = {:rabbit_amqqueue, :emit_unresponsive_local, [vhost, queue_timeout]}
+      all_mfa  = {:rabbit_amqqueue, :emit_unresponsive, [nodes, vhost, queue_timeout]}
+      {chunks, mfas} = case local_opt do
+          true  -> {1, [local_mfa]};
+          false -> {Kernel.length(nodes), [all_mfa]}
+        end
+        RpcStream.receive_list_items(node_name, mfas, timeout, [:name], chunks)
```
hairyhum: Can we have more info items from the queue?

hairyhum: Can we have more info items?

lib/rabbitmq/cli/ctl/commands/list_unresponsive_queues_command.ex
hairyhum: `queue_timeout` should be in milliseconds, which is inconsistent with `timeout` flag for other command, which is in seconds.

hairyhum:
Should we also address `list_queues` command?

lib/rabbitmq/cli/ctl/commands/list_unresponsive_queues_command.ex
michaelklishin: That's a good point.

michaelklishin:
@hairyhum address `list_queues` how?

lib/rabbitmq/cli/ctl/commands/list_unresponsive_queues_command.ex
dcorbacho: Right, we can add the same than on `down` queues

lib/rabbitmq/cli/ctl/commands/list_unresponsive_queues_command.ex
```diff
@@ -25,35 +27,50 @@ defmodule RabbitMQ.CLI.Ctl.Commands.ListUnresponsiveQueuesCommand do
...
   end
+  def merge_defaults([], opts) do
+      merge_defaults(~w(name), opts)
+  end
 
   def switches(), do: [queue_timeout: :integer, local: :boolean]
 
   def usage() do
-    "list_unresponsive_queues [--local] [--queue-timeout <queue-timeout>]"
+    "list_unresponsive_queues [--local] [--queue-timeout <queue-timeout>] [<queueinfoitem> ...]"
```
hairyhum: Info items for this command are not the same as `queueinfoitem` for the `list_queues` command and they are validated (so not all `queueinfoitem` can be used with this command). When printing `rabbitmqctl help` this can be confusing.

PR #218, opened 2017-08-07
Set a timeout on list_queues and a warning if unresponsive ones are found

dcorbacho:
Related to rabbitmq-cli#207

### DIFF
--- a/lib/rabbitmq/cli/ctl/commands/list_queues_command.ex
+++ b/lib/rabbitmq/cli/ctl/commands/list_queues_command.ex
@@ -83,12 +83,16 @@ defmodule RabbitMQ.CLI.Ctl.Commands.ListQueuesCommand do
           {_, false, true}  -> {Kernel.length(nodes), [online_mfa]};
           {_, true, false}  -> {1, [offline_mfa]}
         end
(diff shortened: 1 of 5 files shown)


lib/rabbitmq/cli/ctl/commands/list_queues_command.ex
```diff
@@ -83,12 +83,16 @@ defmodule RabbitMQ.CLI.Ctl.Commands.ListQueuesCommand do
...
+          fn({{:error, {:badrpc, {:timeout, to}}}, :finished}) ->
+            {{:error, {:badrpc, {:timeout, to, "Some queue(s) are unresponsive, use list_unresponsive_queues command."}}}, :finished};
+            (any) -> any
+          end)
       end)
   end
 
   defp default_opts() do
-    %{vhost: "/", offline: false, online: false, local: false}
+    %{vhost: "/", offline: false, online: false, local: false, timeout: 60}
```
hairyhum: Timeout is parsed in `rabbitmqctl.ex` and passed to the command as `infinity`. The default value should override `infinity`. See `node_health_check_command.ex`

PR #229, opened 2017-11-28
Adapt rabbitmq-diagnostics memory_breakdown to server master

### LINKED ISSUE #228: rabbitmq-diagnostics memory_breakdown fails with an exception
Most likely due to the changes in https://github.com/rabbitmq/rabbitmq-server/pull/1404.

michaelklishin:
Specifically to https://github.com/rabbitmq/rabbitmq-server/pull/1404.

Closes #228.

[#153217000]

### DIFF
--- a/lib/rabbitmq/cli/diagnostics/commands/memory_breakdown_command.ex
+++ b/lib/rabbitmq/cli/diagnostics/commands/memory_breakdown_command.ex
@@ -83,14 +83,14 @@ defmodule RabbitMQ.CLI.Diagnostics.Commands.MemoryBreakdownCommand do
   defp compute_relative_values(all_pairs) do
     strategy  = Keyword.get(all_pairs, :strategy)
     num_pairs = Keyword.delete(all_pairs, :strategy)
-    # Includes RSS, allocated and runtime-used ("erlang") values
+    # Includes RSS, allocated and runtime-used ("erlang") values.
     # See https://github.com/rabbitmq/rabbitmq-server/pull/1404.
     totals    = Keyword.get(num_pairs, :total)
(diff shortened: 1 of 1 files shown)


lib/rabbitmq/cli/diagnostics/commands/memory_breakdown_command.ex
```diff
@@ -81,8 +81,19 @@ defmodule RabbitMQ.CLI.Diagnostics.Commands.MemoryBreakdownCommand do
...
   defp compute_relative_values(all_pairs) do
-    pairs = Keyword.delete(all_pairs, :total)
-    total = Enum.reduce(pairs, 0, fn({_, bytes}, acc) -> acc + bytes end)
+    strategy  = Keyword.get(all_pairs, :strategy)
+    num_pairs = Keyword.delete(all_pairs, :strategy)
+    # Includes RSS, allocated and runtime-used ("erlang") values.
+    # See https://github.com/rabbitmq/rabbitmq-server/pull/1404.
+    totals    = Keyword.get(num_pairs, :total)
+    pairs     = Keyword.delete(num_pairs, :total)
+    total     = Keyword.get(totals, strategy) ||
```
hairyhum: The new memory reporting API reports strategy but does not take it to account when reporting the breakdown. For example if `erlang` strategy is configured, preallocated memory will still be reported.
To create a reasonable percentage breakdown we should not select total using strategy. Instead maximal value of three totals can be used, or use `rss` [like HTTP API](https://github.com/rabbitmq/rabbitmq-management/blob/master/src/rabbit_mgmt_wm_node_memory.erl#L74).

PR #257, opened 2018-08-03
Require Elixir 1.6.6+, adding support for Elixir 1.7.x at the same time

dumbbell:
Elixir 1.6.6 is the first version to support Erlang 21.0. Bumping the minimum to this version eases packaging: packagers don't have to think about Erlang/Elixir combinations which work.

The upper limit is set to 1.8.0 (not included) because 1.7.x should work fine. However, 1.8.0 will probably drop Erlang 19.x support, at which point we should review this version pinning.

[#159500280]

### DIFF
--- a/lib/rabbitmq/cli/core/parser.ex
+++ b/lib/rabbitmq/cli/core/parser.ex
@@ -141,7 +141,12 @@ defmodule RabbitMQ.CLI.Core.Parser do
     switches = default_switches()
     aliases = default_aliases()
     {options, tail, invalid} =
-      OptionParser.parse_head(input, strict: switches, aliases: aliases)
+      OptionParser.parse_head(
+        input,
+        strict: switches,
+        aliases: aliases,
+        allow_nonexistent_atoms: true,
+      )
     norm_options = normalize_options(options, switches) |> Map.new
     {norm_options, tail, invalid}
   end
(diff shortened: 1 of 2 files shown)


dumbbell:
The CI pipeline for the `master` branch now tests rabbitmq-cli with Erlang 21.0 and Elixir 1.7.1. It currently fails as expected because `mix.exs` requires Elixir 1.6.x.

dumbbell:
With Elixir 1.7.1, I get a single failure in the testsuite:

```
  * test Auto-completes a command (18.6ms)

  1) test Auto-completes a command (AutoCompleteTest)
     test/core/auto_complete_test.exs:23
     ** (MatchError) no match of right hand side value: []
     code: ["canis_aureus"] = @subject.complete("rabbitmqctl", ["--node", "foo", "--quet", "canis_a"])
     stacktrace:
       test/core/auto_complete_test.exs:29: (test)
```

dumbbell:
The failure is caused by a change of behavior in `OptionParser`. Perhaps related to the following deprecation in Elixir 1.7.0:

```
[OptionParser] Deprecate dynamic option parser mode that depended on atoms to be previously loaded and therefore behaved inconsistently
```

Passing `allow_nonexistent_atoms: true` to `OptionParser restores the previous behavior. I'm testing the patch right now.

dumbbell:
The patch was tested successfully with both Elixir 1.6.6 and 1.7.1.

PR #259, opened 2018-09-19
Quorum queue

kjnilsson:
See: https://github.com/rabbitmq/rabbitmq-server/pull/1706

Includes a new cli tool with commands to manually adjust quorum queue membership.

### DIFF
--- a/lib/rabbitmq/cli/ctl/commands/quorum_status_command.ex
+++ b/lib/rabbitmq/cli/ctl/commands/quorum_status_command.ex
@@ -1,0 +1,56 @@
+## The contents of this file are subject to the Mozilla Public License
+## Version 1.1 (the "License"); you may not use this file except in
+## compliance with the License. You may obtain a copy of the License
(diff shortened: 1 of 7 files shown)


lib/rabbitmq/cli/ctl/commands/quorum_status_command.ex
```diff
@@ -0,0 +1,56 @@
...
+  end
+  def validate([_], _), do: :ok
+  def validate([], _) do
+    {:validation_failure, :not_enough_args}
+  end
+
+  use RabbitMQ.CLI.Core.RequiresRabbitAppRunning
+
+  def run([name] = args, %{node: node_name, vhost: vhost}) do
+    :rabbit_misc.rpc_call(node_name, :rabbit_quorum_queue, :status, [vhost, name])
```
hairyhum: There should be a better error message if trying to get status of a classic queue. Currently there is a case clause.

hairyhum:
Need tests.

PR #263, opened 2018-10-02
New command: rabbitmq-plugins directories

### LINKED ISSUE #261: Introduce rabbitmq-plugins directories
It should be easier to inspect plugin directories (both archives and expanded). `rabbitmq-plugins directories` can be a command that prints them out.

Per discussion with @hairyhum.

michaelklishin:
This introduces a new command, `rabbitmq-plugins directories`, which displays plugin directories and enabled plugins file path.

Closes #261.

[#160792758]

### DIFF
--- a/lib/rabbitmq/cli/plugins/commands/directories_command.ex
+++ b/lib/rabbitmq/cli/plugins/commands/directories_command.ex
@@ -61,7 +61,7 @@ defmodule RabbitMQ.CLI.Plugins.Commands.DirectoriesCommand do
   def usage, do: "directories [--offline] [--online]"
 
   def banner([], %{online: false, offline: true}) do
-    "Listing plugin directories..."
+    "Listing plugin directories inferred from environment..."
   end
 
   def banner([], %{online: true, offline: false, node: node}) do
--- a/lib/rabbitmq/cli/plugins/commands/directories_command.ex
+++ b/lib/rabbitmq/cli/plugins/commands/directories_command.ex
@@ -16,7 +16,7 @@
 
 defmodule RabbitMQ.CLI.Plugins.Commands.DirectoriesCommand do
   alias RabbitMQ.CLI.Plugins.Helpers, as: PluginHelpers
-  alias RabbitMQ.CLI.Core.{Helpers, Validators}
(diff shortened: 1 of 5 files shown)


lib/rabbitmq/cli/plugins/commands/directories_command.ex
```diff
@@ -0,0 +1,119 @@
...
+    :ok
+  end
+
+  def validate_execution_environment(args, %{offline: true} = opts) do
+    Validators.chain([&Helpers.require_rabbit_and_plugins/2,
+                      &PluginHelpers.enabled_plugins_file/2,
+                      &Helpers.plugins_dir/2],
+                     [args, opts])
+  end
+  def validate_execution_environment(args, %{offline: false} = opts) do
```
hairyhum: Is this clause necessary? `validate_execution_environment` happens after `merge_defaults` and `validate`.

lib/rabbitmq/cli/plugins/commands/directories_command.ex
```diff
@@ -0,0 +1,119 @@
...
+    "Listing plugin directories used by node #{node}"
+  end
+
+  def run([], %{online: true, node: node_name}) do
+    do_run fn(key) ->
+      :rabbit_misc.rpc_call(node_name, :rabbit_plugins, key, [])
+    end
+  end
+
+  def run([], %{offline: false, node: node_name}) do
```
hairyhum: Same as the previous comment. This clause should be matched in the previous one.

michaelklishin:
@hairyhum ready for another round.

hairyhum:
`this command requires the target node to be running` is a standard error, but it does not mention option to use the `--offline` flag. I don't know if we should address that or not, but it would be nice to see a message mentioning `--offline`.

michaelklishin:
@hairyhum can you please be a bit more specific? What is wrong with it? It uses `rabbitmq_plugins`, which is also used by `RabbitMQ.CLI.Plugins.Helpers`. What should be used instead?

michaelklishin:
@hairyhum ready for another round (together with updated https://github.com/rabbitmq/rabbitmq-server/pull/1720).

PR #272, opened 2018-11-01
Make stop idempotent using --idempotent flag

dcorbacho:
[#161309687]


hairyhum:
Can we have a test like this at least?
https://github.com/rabbitmq/rabbitmq-cli/blob/master/test/stop_command_test.exs#L48

PR #282, opened 2018-12-07
New diagnostics command: runtime_thread_stats

michaelklishin:
Uses Erlang's microstate accounting data to produce
high level thread state/activity breakdown.

Usage:

``` shell
# sample for 5 seconds
rabbitmq-diagnostics runtime_thread_stats -i 5
```

Per discussion with @gerhard.

Depends on https://github.com/rabbitmq/rabbitmq-common/pull/286.

### DIFF
--- a/lib/rabbitmq/cli/diagnostics/commands/discover_peers_command.ex
+++ b/lib/rabbitmq/cli/diagnostics/commands/discover_peers_command.ex
@@ -25,7 +25,9 @@ defmodule RabbitMQ.CLI.Diagnostics.Commands.DiscoverPeersCommand do
   def validate([_|_], _) do
     {:validation_failure, :too_many_args}
   end
+
   def validate(_, _), do: :ok
+  use RabbitMQ.CLI.Core.RequiresRabbitAppRunning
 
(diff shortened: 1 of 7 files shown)


gerhard:
👍 this looks good, didn't have time to try it out though.

While I remember, it might be worth mentioning in the release notes that if Erlang is compiled with `--with-microstate-accounting=extra`, the command will return more information: https://github.com/erlang/otp/blob/OTP-21.1.4/erts/configure.in#L312-L325

gerhard:
👍 looking good, didn't have time to try it out though.

While I remember, it might be worth mentioning the [`--with-microstate-accounting=extra`](https://github.com/erlang/otp/blob/OTP-21.1.4/erts/configure.in#L312-L325) Erlang/OTP configure flag. [We do this by default in all Erlang packages in the rabbitmq-server-boshrelease](https://github.com/rabbitmq/rabbitmq-server-boshrelease/search?q=--with-microstate-accounting%3Dextra&unscoped_q=--with-microstate-accounting%3Dextra).

michaelklishin:
Backported to `v3.7.x`.

PR #290, opened 2018-12-28
Pad info items in the table formatter to be at least as wide as header.

hairyhum:


michaelklishin:
Backported to `v3.7.x`.

michaelklishin:
I had to revert it after running all integration test suites: now all code that parses table output (bad idea but unfortunately widely done by various deployment automation tools) will have to be adapted to strip trailing whitespace. I'm not convinced it's worth the improvement.

hairyhum:
I'm going to give it another try by making it optional.

PR #291, opened 2019-01-07
Feature flags support + make #amqqueue{} private

dumbbell:
See rabbitmq/rabbitmq-server#1818 for a description of this change.

### DIFF
--- a/lib/rabbit_common/records.ex
+++ b/lib/rabbit_common/records.ex
@@ -17,7 +17,6 @@ defmodule RabbitCommon.Records do
   require Record
   import Record, only: [defrecord: 2, extract: 2]
 
-  defrecord :amqqueue, extract(:amqqueue, from_lib: "rabbit_common/include/rabbit.hrl")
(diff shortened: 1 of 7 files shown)


lib/rabbit_common/records.ex
```diff
@@ -17,7 +17,6 @@ defmodule RabbitCommon.Records do
   require Record
   import Record, only: [defrecord: 2, extract: 2]
 
-  defrecord :amqqueue, extract(:amqqueue, from_lib: "rabbit_common/include/rabbit.hrl")
```
michaelklishin: This is because `amqqueue` is [becoming private](https://github.com/rabbitmq/rabbitmq-common/pull/291) with the introduction of feature flags.

dumbbell: Is it a question or do you answer another comment I didn't see?

michaelklishin: This is for those who are not sure why this record is no longer imported here.

dumbbell: Ok, thank you.

Issue #292, opened 2019-01-14
More/progressive health check commands


michaelklishin:
Yes, even more. Per discussion with @gerhard:

 * There is no single health check command that would be "universal": too many things can go wrong and would be considered a failure by different teams
 * There are node-local and cluster-wide checks, which should be reflected in command names
 * Health checks are stages (just like human or animal health checks), so we need commands that perform increasingly comprehensive checks that will have an increasing likelihood of false positives, e.g. `node_health_check` today checks every channel process which takes a long time with 10s of thousands of channels

Below is a proposal draft that will be refined as we go.

# Introduction

Since relatively few multi-service systems that use messaging can be considered
completely identical and different operators consider different things to be
within the normal parameters, team RabbitMQ (and some other folks who work on data services and their automation [1][2]) has long concluded that
there is no such thing as a "one true way to health check" a RabbitMQ node.

The Docker image maintainer community have arrived at a similar conclusion

Things get even more involved with clusters since distributed system monitoring,
the level of fault tolerance acceptable for a given system,
and preferred ways of reacting/recovering can vary even greatly from
ops team to ops team.

Another important aspect of node monitoring is how it should be altered during
upgrades. This proposal doesn't cover that part.

## Two Types of Health Checks

The proposal is to classify every health check RabbitMQ offers into one of
two categories:

 * Node-local checks
 * Cluster checks

Each category will have a number of checks organized into stages, with
increasingly more aspects of the system checked. This means the probability
of false positives for higher stages will also be higher. Which stage
is used by a given deployment is a choice of that system's operators.

#### Stage 1

What `rabbitmqctl ping` offers today: it ensures that the runtime is running
and (indirectly) that CLI tools can authenticate to it.

This is the most basic check possible. Except for the CLI tool authentication
part, the probability of false positives can be considered approaching 0
except for upgrades and maintenance windows.

#### Stage 2

Includes all checks in stage 1 plus makes sure that `rabbitmqctl status`
(well, the function that backs it) succeeds.

This is a common way of sanity checking a node.
The probability of false positives can be considered approaching 0
except for upgrades and maintenance windows.

#### Stage 3

Includes all checks in stage 2 plus checks that the RabbitMQ application is running
(not stopped/"paused" with `rabbitmqctl stop_app` or the Pause Minority partition
handling strategy) and there are no resource alarms.

The probability of false positives is generally low but during upgrades and
maintenance windows can raise significantly.

[...]

PR #294, opened 2019-01-20
Introoduce rabbitmq-diagnostics is_running, is_booting

michaelklishin:
Part one of #292 introduces

 * `rabbitmq-diagnostics is_running` which returns a success if the node is fully booted and running (as reported by `rabbit:is_booted/1`). This command can be used as/in a health check and meant to be used both interactively and non-interactively.

 *  * `rabbitmq-diagnostics is_booting` which returns a success if the node is currently booting (as reported by `rabbit:is_booting/1`). For full booted (running) and stopped nodes it would report a failure. This command is meant for interactive use only and mostly for completeness/troubleshooting. It is *NOT* a good fit for health checks since it will return a failure the vast majority of the time (unless you are lucky to catch a node booting or booting takes a long time)


michaelklishin:
This is a pretty trivial change => will self-QA to continue with #292.

PR #296, opened 2019-01-21
Introduce `rabbitmq-diagnostics alarms`

michaelklishin:
Part of #292.

When QA'ing, it is helpful to set some alarms on the node with

```
rabbitmqctl set_disk_free_limit "10000GB"; rabbitmqctl set_vm_memory_high_watermark absolute 100kB
```

and clean them with

```
rabbitmqctl set_disk_free_limit "10GB"; rabbitmqctl set_vm_memory_high_watermark 0.4
```


michaelklishin:
Merging per discussion with @gerhard.

PR #297, opened 2019-01-21
Add active and activity status fields to list_consumers

acogoluegnes:
[#163298456]

References rabbitmq/rabbitmq-management#1838


lib/rabbitmq/cli/ctl/commands/list_consumers_command.ex
```diff
@@ -27,12 +27,12 @@ defmodule RabbitMQ.CLI.Ctl.Commands.ListConsumersCommand do
...
 
   @info_keys ~w(queue_name channel_pid consumer_tag
-                ack_required prefetch_count single_active arguments)a
+                ack_required prefetch_count active activity_status arguments)a
 
   def info_keys(), do: @info_keys
 
   def merge_defaults([], opts) do
-    {Enum.map(@info_keys -- [:single_active], &Atom.to_string/1), Map.merge(%{vhost: "/", table_headers: true}, opts)}
+    {Enum.map(@info_keys -- [:active, :activity_status], &Atom.to_string/1), Map.merge(%{vhost: "/", table_headers: true}, opts)}
```
hairyhum: Should activity be in the default arguments?

acogoluegnes: Not sure. Those 2 fields matter when single active consumer is on or when investigating an issue with quorum queues. To me these are specific cases, not the most common ones, that's why the 2 fields are not in the default ones. I was also thinking that adding them (or one) would bring up questions for cases where they don't matter. What do you think @kjnilsson?

michaelklishin: If most users won't need them I'd say let's leave them out.

michaelklishin:
@acogoluegnes I rebased this on top of `master` and force-pushed, please take a look.

lib/rabbitmq/cli/ctl/commands/list_consumers_command.ex
hairyhum: Yes, the `activity_status` field is only needed in case of investigations, but the `active` field may be useful in general to detect if there is a need for investigation.

acogoluegnes: This is reasonable.

acogoluegnes: OK, the active field is now in the defaults.

PR #301, opened 2019-01-23
Make it possible for formatters to define arguments.

hairyhum:
Add new callbacks to formatter_behavior: switches/0 and aliases/0
Merge them with command switches and aliases on parsing.

There is an example switch in the table formatter, enabling header padding.


michaelklishin:
It's an interesting idea, I like where this is headed.

PR #306, opened 2019-01-30
Make it possible to get command help by typing 'rabbitmqctl mycommand --help'

hairyhum:
Add a new default switch `help`
Add an alias `?`

Make rabbitmqctl display command help if this option is true for known commands.


michaelklishin:
Very nice 👍

PR #308, opened 2019-02-01
Support bare --help

### LINKED ISSUE #307: --help should exit with the code of 0
It currently returns 64 ("invalid arguments").

The team's consensus around this is the following. Exiting with an `EX_USAGE` makes sense for no argument invocations (e.g. `rabbitmqctl`). `curl` and `grep` use a non-zero exit code in that case. When the user requests `--help` and the tool displays it to her, it makes sense to exit with the code of 0.

michaelklishin:
Note that there's a different between the case with no arguments
(exits with EX_USAGE) and --help (exits successfully).

Per discussion with the team.

Closes #307.

### DIFF
--- a/lib/rabbitmqctl.ex
+++ b/lib/rabbitmqctl.ex
@@ -46,6 +46,22 @@ defmodule RabbitMQCtl do
     |> handle_shutdown
   end
 
+  def exec_command([] = unparsed_command, _) do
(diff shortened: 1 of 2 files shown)


michaelklishin:
`rabbitmqctl_integration_SUITE` passes, backported to `v3.7.x`

PR #313, opened 2019-02-12
Use elixir internal distance function instead of a library for command similarity comparison.

hairyhum:
Elixir provides String.jaro_distance function, which is similar to
the levenstain distance.
This allows us to remove a dependency.


michaelklishin:
From my testing it works just as well as the previous implementation but with one fewer dependency 👍👍

michaelklishin:
Backported to `v3.7.x`.

PR #316, opened 2019-02-14
Help command output revision

hairyhum:
`rabbitmqctl help` prints a lot of info which is not that helpful. 
This PR changes the output of the command to be more like `git help`:
```
./scripts/rabbitmqctl help

Usage:
rabbitmqctl [-n <node>] [-t <timeout>] [-l|--longnames] [-q|--quiet] <command> [<command options>]

Commands:
Print this help and commad specific help:
    help                          
Application management:
    await_startup                 Waits for the RabbitMQ application to start on existing node
    ping                          Checks that the node process is running nad registered with EPMD
    reset                         Instructs a RabbitMQ node to leave the cluster and eturn to its virgin state
    rotate_logs                   Instructs the RabbitMQ node to perform internal log rotation
    shutdown                      Stops the Erlang node on which RabbitMQ is running and waits for the OS process to exit
    start_app                     Starts the RabbitMQ application inside the Erlang VM node
    stop                          Stops the Erlang node on which RabbitMQ is running
    stop_app                      Stops the RabbitMQ application, leaving the runtme (Erlang VM) running
    wait                          Waits for the RabbitMQ application to start
```

There are two new optional callbacks in the command behaviour:
`help_section/0` - atom tag to categorize commands
`description/0` - String (binary) description of the command. Should be relatively short.

TODO: more descriptiotns and help sections for plugin and diagnostics commands.


hairyhum:
I tried to make sections closer to what we have in https://next.rabbitmq.com/rabbitmqctl.8.html
@michaelklishin do you want to modify the mandoc to be more like the help message instead?

michaelklishin:
@hairyhum yes, I think the source of truth here should be the CLI codebase. Happy to modify man pages after this is in.

PR #317, opened 2019-02-19
Add shrink command to rabbitmq-queues

kjnilsson:
This command shrinks all quorum queue clusters with members for a provided rabbit node and returns a table of the result.

Requires: https://github.com/rabbitmq/rabbitmq-server/pull/1890


kjnilsson:
replaced with #318

PR #318, opened 2019-02-20
rabbitmq-queues grow and shrink commands

kjnilsson:
Supersedes #317 

Requires: https://github.com/rabbitmq/rabbitmq-server/pull/1894


michaelklishin:
One thing I've noticed while QA'ing this is that when there are no results (e.g. because the node provided already has replicas for all matching queues), the command simply produces an empty line. However that's a general issue for anything that uses the table formatter.

PR #321, opened 2019-02-21
New command that runs full sweep GC for all processes on the target node

### LINKED ISSUE #319: Introduce a command that forces fullsweep GC on all processes on a node
Occasionally it is useful to have due to the lazy nature of Erlang GC.

This command was originally developed as a live coding session exercise but we decided to keep it.

michaelklishin:
Based on and includes #316, closes #319.

This was collectively reviewed during a live coding session.


michaelklishin:
The new command was backported to `v3.7.x` (but not #316).

PR #322, opened 2019-02-24
Spelling

jsoref:
Generated by https://github.com/jsoref/spelling `f`; to maintain your repo, please consider `fchurn`

Note: I've filed some issues and made some PRs on some of the pkg_ items changed below, so they may technically not be incorrect today, but I believe that the changes should be made both here and upstream.


michaelklishin:
Thank you!

PR #324, opened 2019-02-26
ctl forget_cluster_node: improve error formatting

### LINKED ISSUE #323: forget_cluster_node should produce a more helpful error message when the node that's being removed is running
In a cluster of A, B and C, if the RabbitMQ app is running on node B and the operator attempts to remove it with

```
rabbitmqctl forget_cluster_node -n rabbit@host-a rabbit@host-b
```

then the operation will fail with an error tuple dumped without any formatting.

michaelklishin:
Fixes #323.

To reproduce:

 * Form a cluster of 2 or more nodes
 * Try to remove a node where the `rabbit` app is still running with `rabbitmqctl forget_cluster_node -n rabbit@host-a rabbit@host-b` or similar
 * Observe a sensible error with a hint as to what the operator should do


michaelklishin:
Backported to `v3.7.x`.

PR #328, opened 2019-03-06
ctl shutdown: infer hostnames from node names

### LINKED ISSUE #327: `ctl shutdown`: use` erlang:node/0` when detecting if target node is remote
As we concluded it is a more reliable source. `inet_db:hostname/0` can return the same value on all cluster nodes as it doesn't account for node names passed from the command line nor `ERL_INETRC`.

One such environment is `cf-rabbitmq-release`, as was [discovered](https://github.com/rabbitmq/rabbitmq-cli/pull/309#issuecomment-470102112) by @nodo and @gerhard.

michaelklishin:
inet_db is not a very reliable source as it doesn't take
node name CLI arguments and ERL_INETRC file settings.
That can lead to false positives in environments where
inet_db returns the same value (e.g. `localhost`) for
every cluster member.

Per discussion with @gerhard.

Closes #327.
References #309.

### DIFF
--- a/lib/rabbitmq/cli/core/node_name.ex
+++ b/lib/rabbitmq/cli/core/node_name.ex
@@ -34,6 +34,40 @@ defmodule RabbitMQ.CLI.Core.NodeName do
   def hostname, do: :inet_db.gethostname() |> List.to_string()
 
   @doc """
+  Get hostname part of current node name
+  """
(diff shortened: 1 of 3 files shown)


michaelklishin:
👍

michaelklishin:
Backported to `v3.7.x`.

PR #344, opened 2019-04-18
Feature flag commands: handle missing rabbit_ff_extra modules on target node

michaelklishin:
This makes `ctl list_feature_flags` handle missing rabbit_ff_extra modules on target node.

@dumbbell @lukebakken @gerhard  if this is the behavior we want I will extend it to more commands. Alternatively we could return an empty list of feature flags (this may be more operator-friendly during upgrades).

[#163980302]


dumbbell:
Yes, it should return an empty list of feature flags: RabbitMQ itself does exactly that, i.e. the absence of the feature flags subsystem is not an error.

michaelklishin:
Ready for another round.

michaelklishin:
Merging per verbal approval from @dumbbell.

PR #347, opened 2019-04-24
Add a flag to cipher_suites command to get all ciphers. Add --format flag to select format. Remove openssl_format flag

hairyhum:
Support new map format for cipher suites added in OTP-20.3
Support `--all` flag to get all available ciphers

Addresses #342
Depends on changes in rabbitmq-server https://github.com/rabbitmq/rabbitmq-server/pull/1991

### DIFF
--- a/lib/rabbitmq/cli/diagnostics/commands/cipher_suites_command.ex
+++ b/lib/rabbitmq/cli/diagnostics/commands/cipher_suites_command.ex
@@ -17,24 +17,38 @@ defmodule RabbitMQ.CLI.Diagnostics.Commands.CipherSuitesCommand do
   @behaviour RabbitMQ.CLI.CommandBehaviour
   use RabbitMQ.CLI.DefaultOutput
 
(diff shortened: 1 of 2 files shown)


hairyhum:
It feels like formats would be cleaner if there is a single string flag e.g. `--format=openssl`
We may leave `--openssl-format || --erlang-format` options and remove them in 3.8

michaelklishin:
We can keep them and make them delegate to `--format=…` ;)

hairyhum:
I'm not sure how keeping them and adding a new flag makes things better.

michaelklishin:
Only for backwards compatibility. We will remove them from the docs.

hairyhum:
By the way, `erlang_format` was only added in master and `openssl_format` was made default in master. I'm not sure if that's on purpose, but I'm leaning towards moving to `--format` flag and removing `--erlang-format` before it's released.

hairyhum:
I'm going to submit another PR for 3.7

michaelklishin:
The default change for 3.8 was intentional.

PR #350, opened 2019-04-25
diagnostics erlang_version: support --offline

### LINKED ISSUE #315: `rabbitmq-diagnostics erlang_version --offline`
To display the Erlang version used locally. Note that retrieving full version from OTP is remarkably painful so we have to use a helper function RabbitMQ already has: `rabbit_misc:otp_release/0`.

michaelklishin:
Closes #315.


gerhard:
I was in the mid-review when @dcorbacho beat me to it.

I was going to suggest:

```
CLI Erlang/OTP version ...
```

As an alternative to:

```
Will print local Erlang/OTP version ...
```

michaelklishin:
@gerhard feel free to change the wording and push directly to master/v3.7.x. Thank you.

PR #352, opened 2019-05-07
WIP Make rabbitmqctl work with any endianness.

hairyhum:
Elixir checks endianness and prints this warning:
```
Elixir is running in a system with a different endianness than the one its source code was compiled in. Please make sure Elixir and all source files were compiled in a machine with the same endianness as the current one: big"
```
on big endian systems.
This warning is mostly there because of precompiled regexp fearure in Elixir.

To make code work on all systems, regexes are now recompiled in runtime.

Another question is: should we remove the warning now when the code may run on any endianness?

### DIFF
--- a/lib/rabbitmq/cli/core/listeners.ex
+++ b/lib/rabbitmq/cli/core/listeners.ex
@@ -179,7 +179,8 @@ defmodule RabbitMQ.CLI.Core.Listeners do
     # This simplistic way of distinguishing IPv6 addresses,
     # networks address ranges, etc actually works better
     # for the kind of values we can get here than :inet functions. MK.
-    case value =~ ~r/:/ do
+    regex = Regex.recompile!(~r/:/)
+    case value =~ regex do
       true -> "[#{value}]"
       false -> value
     end
--- a/lib/rabbitmq/cli/ctl/commands/help_command.ex
+++ b/lib/rabbitmq/cli/ctl/commands/help_command.ex
@@ -290,8 +290,8 @@ short            | long          | description
     end
   end
 
-  defp strip_rabbitmq_prefix(value) do
-    Regex.replace(~r/^rabbitmq_/, value, "")
+  defp strip_rabbitmq_prefix(value, regex) do
+    Regex.replace(regex, value, "")
   end
 
   defp format_known_plugin_name_fragments(value) do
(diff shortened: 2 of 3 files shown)


dumbbell:
I tested this patch like this:
1. Created a generic-unix package from the `master` branch on `amd64`.
2. Started RabbitMQ from that package on `s390x`.
3. Executed `rabbitmqctl status` and got:

    ```
    ./sbin/rabbitmqctl status
    warning: Elixir is running in a system with a different endianness than the one its source code was compiled in. Please make sure Elixir and all source files were compiled in a machine with the same endianness as the current one: big
    Status of node rabbit@s390x ...
        (rabbitmqctl) lib/rabbitmqctl.ex:106: RabbitMQCtl.exec_command/2
        (rabbitmqctl) lib/rabbitmqctl.ex:45: RabbitMQCtl.main/1
        (elixir) lib/kernel/cli.ex:105: anonymous fn/3 in Kernel.CLI.exec_fun/2
    ```

4. Recompiled the generic-unix package from the `endian_insensitive` branch on `amd64`.
5. Restarted RabbitMQ from that new package on `s390x`.
6. Executed `rabbitmqctl status` and got:

    ```
    ./sbin/rabbitmqctl status
    warning: Elixir is running in a system with a different endianness than the one its source code was compiled in. Please make sure Elixir and all source files were compiled in a machine with the same endianness as the current one: big
    Status of node rabbit@s390x ...
    Runtime

    OS PID: 2790
    OS: Linux
    Uptime (seconds): 444
    RabbitMQ version: 3.8.0+beta.3.13.g51946c2
    Node name: rabbit@s390x
    Erlang configuration: Erlang/OTP 21 [erts-10.3.4] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:64]
    Erlang processes: 246 used, 1048576 limit
    Scheduler run queue: 1
    Cluster heartbeat timeout (net_ticktime): 60
    ...
    ```

Therefore, rabbtimqctl(8) looks to work after the patch. However, we still get the warning. Is this expected?

hairyhum:
There is no change to suppress the warning. I'm not sure if that is possible without changes to the elixir core, because it calls `io:format` there.

michaelklishin:
The warning is benign assuming that CLI tools are functional with this change but we definitely should contribute a change to Elixir core.

hairyhum:
Another PR to elixir: https://github.com/elixir-lang/elixir/pull/9040

dumbbell:
Thank you! Merged to `master` and backported to `v3.7.x`.

PR #353, opened 2019-05-10
Observer CLI integration

michaelklishin:
This introduces a new command, `rabbitmq-diagnostics observer`, which provides a way to start [observer CLI](https://github.com/zhongwencool/observer_cli) on the target node.

Depends on https://github.com/rabbitmq/rabbitmq-server/pull/2006.

Per discussion with @gerhard.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,9 @@
 PROJECT = rabbitmq_cli
 
+dep_observer_cli = git https://github.com/michaelklishin/observer_cli recon-2.4.0
+
 BUILD_DEPS = rabbit_common
+DEPS = observer_cli
 TEST_DEPS = amqp_client rabbit
 
 DEP_EARLY_PLUGINS = rabbit_common/mk/rabbitmq-early-plugin.mk
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 PROJECT = rabbitmq_cli
 
-dep_observer_cli = hex observer_cli 1.4.4
+dep_observer_cli = git https://github.com/zhongwencool/observer_cli 1.4.4
 
 BUILD_DEPS = rabbit_common
 DEPS = observer_cli
--- a/Makefile
+++ b/Makefile
@@ -1,6 +1,6 @@
 PROJECT = rabbitmq_cli
 
-dep_observer_cli = hex observer_cli 1.4.3
+dep_observer_cli = hex observer_cli 1.4.4
 
 BUILD_DEPS = rabbit_common
 DEPS = observer_cli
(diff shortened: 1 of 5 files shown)


gerhard:
This is so cool! Looking forward to QA 💯

michaelklishin:
What the interface looks like:

observer_cli

It is fully controlled with lettered shortcuts + `RET`/`[Enter]` and mimics the GUI Observer Erlang app.

michaelklishin:
Note: one test is disabled as `observer_cli:start/2` currently [does not communicate errors correctly](https://github.com/zhongwencool/observer_cli/pull/54).

zhongwencool:
@michaelklishin `1.4.4` is ready  :0

michaelklishin:
@gerhard this is ready for QA 🙏

gerhard:
Nice, will check it out shortly!

gerhard:
observer_cli is failing to install for me:

```shell
ls $PWD/.erlang.mk
ls: /Users/gerhard/github.com/rabbitmq/20190501/deps/rabbit/.erlang.mk: No such file or directory

ls $PWD/../observer_cli
ls: /Users/gerhard/github.com/rabbitmq/20190501/deps/rabbit/../observer_cli: No such file or directory

gmake deps
 DEP    observer_cli
tar: Error opening archive: Failed to open '/Users/gerhard/github.com/rabbitmq/20190501/deps/rabbit/.erlang.mk/hex/observer_cli.tar'

ls ../observer_cli
cat ../observer_cli/Makefile
noop:
```

umbrella is clean, I have the `observer_cli` branches checked out in both `rabbit` & `rabbitmq_cli`. Will continue with this tomorrow.

zhongwencool:
The diff between `escript` and `remsh` is because `io:rows()` can't get  row in escript, so use a DEFAULT_ROW(Suppose the screen is only 13 inch).
https://github.com/zhongwencool/observer_cli/blob/master/src/observer_cli_lib.erl#L237-L241

I can't find a better way to solve this problem right now.

gerhard:
> The diff between `escript` and `remsh` is because `io:rows()` can't get row in escript, so use a DEFAULT_ROW(Suppose the screen is only 13 inch).

Thanks! If @dumbbell doesn't have a suggestion, happy to leave this as is.

michaelklishin:
> I also think that we should add rabbitmqctl observe in the Monitoring, observability and health checks section

It already is for `rabbitmq-diagnostics`. We no longer add new commands to `rabbitmqctl` unless there are good reason to do so (e.g. it does not clearly fit into any other tool/scope).

michaelklishin:
@gerhard thank you. This currently will fail in our pipelines as we need https://github.com/zhongwencool/observer_cli/pull/57 to go in before Mix can build it on Concourse (due to a Mix behavior neither @dumbbell nor I can explain at the moment). So we will have to do another `rabbitmq-components.mk` update.

PR #354, opened 2019-05-22
remove unsupported type argument from list queues command

luos:
Type is a new argument for 3.8.x. Listing it here causes rabbitmqctl report to
crash with {nocatch,{bad_argument,type}} when it tries to list queues with all
arguments.


michaelklishin:
FTR, it was a backport side effect I haven't noticed (in fb7f691a14d2010aeb0f7098ba6bca74db8875d2).

michaelklishin:
Thank you!

PR #358, opened 2019-06-13
Command to log schema information

### LINKED ISSUE #330: rabbitmq-diagnostics command to log schema information
It can be useful when troubleshooting an upgrade. Note that we don't have to log every gory detail of the schema. Tables and their properties, replica information, schema cookie/"epoch" should be enough for starters.

dcorbacho:
Logs by default name, cookie, active_replicas and user_properties
Any other mnesia property can be requested as info key

Needs https://github.com/rabbitmq/rabbitmq-server/pull/2033
Closes #330 

[#164550205]


michaelklishin:
Backported to `v3.7.x`.

Issue #325, opened 2019-02-26
rabbitmq-diagnostics command to retrieve N most recent log lines 


michaelklishin:
In many cases it would be convenient to retrieve logs via `rabbitmq-diagnostics` instead of looking for a log file location first. It also can be done remotely.

Related to #299.

gerhard:
This is a great idea, love it!

michaelklishin:
Backported to `v3.7.x`.

PR #360, opened 2019-06-27
Diagnostics commands that access server logs

hairyhum:
Added new commands:

- `log_location` - get default log location
- `log_location --all` - get all logs locations
- `log_tail [-N <number>]` - get last N lines from the default log
- `log_tail_stream [-d <duration>]` - stream log messages from the deafault log

Default N is 10, default duration is `:infinity`.

Log streaming does not detect the log file to be deleted (because it's a hard thig to do), but that's similar to `tail -f`.
Added a new printer, which prints to the std_io, but does not split lines. This is because log data may contain partial lines.

Reading last N lines reads a buffer of data from the end of the file and counts lines. If there are not enough lines - the buffer is increased. This works well enough for reasonable number of lines.

Open questions:

- Should duration be some number of seconds by default instead of `:infinity`?
- Should there be a default number of lines? Is 10 a good number?

Additional features:

- Should there be an argument to ask for entire file?
- Should there be a command to get the tail AND stream the file from the last line returned (like `tail -n 100 -f`)?
- Should it be possible to select non-default log files?

### DIFF
--- a/lib/rabbitmq/cli/command_behaviour.ex
+++ b/lib/rabbitmq/cli/command_behaviour.ex
@@ -34,6 +34,7 @@ defmodule RabbitMQ.CLI.CommandBehaviour do
               | {:error, RabbitMQ.CLI.Core.ExitCodes.exit_code(), [String.t()]}
 
   @optional_callbacks formatter: 0,
+                      printer: 0,
                       scopes: 0,
                       usage_additional: 0,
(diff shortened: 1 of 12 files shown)


hairyhum:
Addresses #325 #326

hairyhum:
Requires https://github.com/rabbitmq/rabbitmq-server/pull/2047

PR #368, opened 2019-07-18
Do not require rabbit app for rabbitmq-diagnostics observer

gerhard:
It works perfectly fine without it. We are more likely to want to look into the Erlang VM if the rabbit app is not running.

Didn't think it was worth writing a test for this, but I am submitting it as a PR in case you think differently @michaelklishin.


michaelklishin:
Backported to `v3.7.x`.

PR #371, opened 2019-08-14
Use new 'description' and 'tag' items on list vhosts

dcorbacho:
Part of https://github.com/rabbitmq/rabbitmq-server/issues/2012

### DIFF
--- a/lib/rabbitmq/cli/ctl/commands/add_vhost_command.ex
+++ b/lib/rabbitmq/cli/ctl/commands/add_vhost_command.ex
@@ -29,7 +29,7 @@ defmodule RabbitMQ.CLI.Ctl.Commands.AddVhostCommand do
   use RabbitMQ.CLI.Core.RequiresRabbitAppRunning
 
   def run([vhost], %{node: node_name, description: desc, tags: tags}) do
-    :rabbit_misc.rpc_call(node_name, :rabbit_vhost, :add, [vhost, desc, tags, Helpers.cli_acting_user()])
+    :rabbit_misc.rpc_call(node_name, :rabbit_vhost, :add, [vhost, desc, parse_tags(tags), Helpers.cli_acting_user()])
   end
   def run([vhost], %{node: node_name}) do
     :rabbit_misc.rpc_call(node_name, :rabbit_vhost, :add, [vhost, Helpers.cli_acting_user()])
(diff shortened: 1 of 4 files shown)


lib/rabbitmq/cli/ctl/commands/add_vhost_command.ex
```diff
@@ -47,4 +58,14 @@ defmodule RabbitMQ.CLI.Ctl.Commands.AddVhostCommand do
...
   def banner([vhost], _), do: "Adding vhost \"#{vhost}\" ..."
+
+  #
+  # Implementation
+  #
+
+  def parse_tags(tags) do
+    String.split(tags, ",", trim: true)
+    |> Enum.map(&String.trim/1)
+    |> Enum.map(&String.to_atom/1)
```
dumbbell: I suppose we can use the existing function in rabbitmq-server

dumbbell: Also same comment as the one attached the other copy of `parse_tags()`: do we really need atoms here?

lib/rabbitmq/cli/ctl/commands/add_vhost_command.ex
```diff
@@ -18,21 +18,32 @@ defmodule RabbitMQ.CLI.Ctl.Commands.AddVhostCommand do
...
   use RabbitMQ.CLI.DefaultOutput
 
-  def usage, do: "add_vhost <vhost>"
+  def usage, do: "add_vhost <vhost> [--description <description> --tags \"<tag1>,<tag2>,<...>\"]"
 
   def usage_additional() do
     [
-      ["<vhost>", "Virtual host name"]
+      ["<vhost>", "Virtual host name"],
+      ["--description <descriptioon>", "Virtual host description"],
```
dumbbell: There is typo in `descriptioon`.

lib/rabbitmq/cli/ctl/commands/add_vhost_command.ex
```diff
@@ -18,21 +18,32 @@ defmodule RabbitMQ.CLI.Ctl.Commands.AddVhostCommand do
...
+      ["--tags <tags>", "Coommand separated list of tags"]
```
dumbbell: There is a typo in `Coommand`. I suppose you mean `Comma` here.

lib/rabbitmq/cli/ctl/commands/add_vhost_command.ex
michaelklishin: We agreed to use atoms in rabbitmq/rabbitmq-server#2080. CLI tools try to depend as little as possible on the server modules and when it comes to working with strings, I'd take Elixir over Erlang any day.

PR #390, opened 2019-12-05
Introduce `rabbitmq-queues check_if_node_is_quorum_critical'

michaelklishin:
Mob: @kjnilsson and several other Pivots.

Part of #389.


michaelklishin:
Will re-submit with a different branch name.

PR #391, opened 2019-12-12
Introduce `rabbitmq-queues check_if_node_is_quorum_critical`

michaelklishin:
Introduce `rabbitmq-queues check_if_node_is_quorum_critical` that returns a non-zero status if shutdown or removal of target node would make one or more quorum queues from sustaining a quorum.

Part one of rabbitmq/rabbitmq-cli#389.


michaelklishin:
Backported to `v3.8.x`.

PR #400, opened 2020-02-28
wait_command.ex: Change `wait` internal sleep to one second

dumbbell:
... down from 10% of the configured timeout.

This has a significant impact on the time it takes to start RabbitMQ in all our testsuites. rabbitmq-ct-helpers sets a wait timeout of 180 seconds. Thus before this patch, the wait loop would sleep for 18
seconds between each check. Given it takes about 1.5 seconds to start RabbitMQ, a lot of time is wasted here.

Here are some numbers after running testsuites with and without this patch:

* `make ct-fast` in rabbitmq-server: 8m15s down to 4m58s
* `make ct` in rabbitmq-mqtt: 9m23s down to 6m43s
* `make ct` in rabbitmq-stomp: 4m31s down to 2m04s


michaelklishin:
This is a great find.

PR #407, opened 2020-04-16
Handle the move of the enabled_plugins file

pjk25:
We look for the enabled_plugins file at the deprecated location, and at the new location. If it exists only at the deprecated location, we attempt to use that file. If the deprecated location is read-only, we allow the cli command to fail (or not) as it always used to.

Part of the changes intended to address https://github.com/rabbitmq/rabbitmq-server/issues/2234

### DIFF
--- a/lib/rabbitmq/cli/plugins/plugins_helpers.ex
+++ b/lib/rabbitmq/cli/plugins/plugins_helpers.ex
@@ -70,17 +70,44 @@ defmodule RabbitMQ.CLI.Plugins.Helpers do
     end
   end
 
+  @spec enabled_plugins_file(map()) :: {:ok, charlist()} | {:error, any()}
   def enabled_plugins_file(opts) do
     case Config.get_option(:enabled_plugins_file, opts) do
-      nil -> {:error, :no_plugins_file}
-      file -> {:ok, file}
+      nil ->
+        case :os.type do
+          {:unix, _} ->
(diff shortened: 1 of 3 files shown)


pjk25:
I believe the travis build is failing because the broker needs to be running with the code from https://github.com/rabbitmq/rabbitmq-server/pull/2298
However, I'm not aware of how to change that temporarily for the PR without affecting other things.

michaelklishin:
Yes, this is a fundamental limitation of Travis' "one project = one repository" philosophy.

lib/rabbitmq/cli/plugins/plugins_helpers.ex
```diff
@@ -70,17 +70,44 @@ defmodule RabbitMQ.CLI.Plugins.Helpers do
...
+                              {:ok, Value} ->
+                                Value
+                              _ ->
+                                app_data = System.get_env("APPDATA")
+                                Path.join([app_data, "RabbitMQ"])
+                            end
+            {:ok, Path.join([rabbitmq_base, "enabled_plugins"])}
+        end
+      file ->
+        {:ok, file}
```
dumbbell, line 101: The duplication of the logic between rabbitmq-server and this repository makes me think more and more there is something wrong with that file, but at a deeper level. Otherwise the patches solve the immediate issue.

PR #414, opened 2020-05-15
Add GitHub Actions workflows

pjk25:
Since these are ExUnit rather than ct based, they aren't generated with 'make github-cations'.

Also includes honeycomb.io integration.

Originally the honeycomb formatter was enabled with "mix test --formatter HoneycombFormatter --formatter ExUnit.CLIFormatter", but that doesn't work on Elixir 1.8's version of mix.

### DIFF
--- a/lib/honeycomb_formatter.ex
+++ b/lib/honeycomb_formatter.ex
@@ -4,9 +4,11 @@ defmodule HoneycombFormatter do
   def init(opts) do
     env = System.get_env()
 
+    :application.ensure_all_started(:os_mon)
+
     config = %{
       directory: "/tmp/honeycomb",
-      github_workflow: env["GITHUB_WORKLOW"] || "unknown",
+      github_workflow: env["GITHUB_WORKFLOW"] || "unknown",
       github_run_id: env["GITHUB_RUN_ID"] || "unknown",
       github_repository: env["GITHUB_REPOSITORY"] || "unknown",
       github_sha: env["GITHUB_SHA"] || "unknown",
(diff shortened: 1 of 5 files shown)


lib/honeycomb_formatter.ex
```diff
@@ -0,0 +1,102 @@
...
+    \"base_rmq_ref\":\"#{config[:base_rmq_ref]}\",
+    \"erlang_version\":\"#{config[:erlang_version]}\",
+    \"elixir_version\":\"#{config[:elixir_version]}\",
+    \"otp_release\":\"#{config[:otp_release]}\",
+    \"cpu_topology\":#{config[:cpu_topology_json]},
+    \"schedulers\":#{config[:schedulers]},
+    \"system_architecture\":\"#{config[:system_architecture]}\",
+    \"system_memory_data\":#{config[:system_memory_data_json]},
+    \"suite\":\"#{suite}\",
+    \"testcase\":\"#{quote_json(testcase)}\",
```
pjk25, line 66: I wanted to avoid introducing a json encoder dependency, but if it can be done test only, I think it would be worth it to avoid the partial string escaping code here.

test/test_helper.exs
```diff
@@ -17,7 +17,11 @@ four_hours = 240 * 60 * 1000
 ExUnit.configure(
   exclude: [disabled: true],
   module_load_timeout: four_hours,
-  timeout: four_hours)
+  timeout: four_hours,
+  formatters: case System.get_env("HONEYCOMB_FORMATTER") do
+    "true" -> [HoneycombFormatter, ExUnit.CLIFormatter]
+    _ -> [ExUnit.CLIFormatter]
+  end)
```
pjk25, line 24: Originally the honeycomb formatter was enabled without a code change using "mix test --formatter HoneycombFormatter --formatter ExUnit.CLIFormatter", but that doesn't work on Elixir 1.8's version of mix.

PR #416, opened 2020-05-15
Handle variable case in this test

michaelklishin:
A follow-up to #415.


michaelklishin:
Backported to `v3.8.x`.

PR #417, opened 2020-05-16
Explain

michaelklishin:
Another follow-up PR to #415.


michaelklishin:
Backported to `v3.8.x`.

PR #409, opened 2020-04-21
Introduce a few new rabbitmq-upgrade commands

michaelklishin:
## Proposed Changes

See #408 for the background.

## Types of Changes

- [ ] Bug fix (non-breaking change)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause an observable behavior change in existing systems)
- [ ] Documentation improvements (corrections, new content, etc)
- [ ] Cosmetic change (whitespace, formatting, etc)

## Checklist

- [x] I have read the `CONTRIBUTING.md` document
- [x] I have signed the CA (see https://cla.pivotal.io/sign/rabbitmq)
- [x] All tests pass locally with my changes
- [x] I have added tests that prove my fix is effective or that my feature works
- [ ] I have added necessary documentation (if appropriate)
- [x] Any dependent changes have been merged and published in related repositories

## Further Comments

To QA this, start a three node cluster and declare a quorum queue in it. It will default to three replicas.

Then

 * Run `./sbin/rabbitmq-upgrade await_online_quorum_plus_one -t 30 --formatter=json` which will succeed
 * `./sbin/rabbitmq-upgrade await_online_quorum_plus_one -t 30` which will succeed
 * Stop node two
 * `./sbin/rabbitmq-upgrade await_online_quorum_plus_one -t 10 --formatter=json` which will fail
 * `./sbin/rabbitmq-upgrade await_online_quorum_plus_one -t 10` which will fail
 * Run  * `./sbin/rabbitmq-upgrade await_online_quorum_plus_one -t 60` and start node two again in the background, the command will succeed after a period of waiting

PR #418, opened 2020-05-19
#409 followup

michaelklishin:
This addresses a typo in #409 as well as a couple of default timeouts that were specified incorrectly (assumed the values was in seconds instead of milliseconds).

Spotted by @gerhard in #409.


michaelklishin:
Backported to `v3.8.x`.

PR #423, opened 2020-05-30
List Raft data directory in status

michaelklishin:
This exposes Raft data directory in `rabbitmq-diagnostics status`. Part of rabbitmq/rabbitmq-server#2354.


michaelklishin:
Backported to `v3.8.x`.

PR #425, opened 2020-06-04
Improves error reporting for user and permission-related commands

michaelklishin:
Part of rabbitmq/rabbitmq-server#2363.


michaelklishin:
Backported to `v3.8.x`.

PR #433, opened 2020-06-25
Make 'rabbitmqctl hipe_compile' a no-op

michaelklishin:
HiPE has been deprecated/only partially supported in Erlang 22
and will be removed completely in Erlang 24 next year.

Part of rabbitmq/rabbitmq-server#2392


michaelklishin:
Backported to `v3.8.x`.

PR #436, opened 2020-06-26
rabbitmqctl export_definitions: ensure that parameter values are maps

### LINKED ISSUE #435: CLI definition export incorrectly serialises federation upstreams
"Incorrectly" here means it uses keyword lists instead of maps:

```
Importing definitions in JSON from a file at "/path/to/defs.plugins.cli.json" ...
Error:
Validation failed

Unrecognised terms [[&lt;&lt;"ack-mode"&gt;&gt;,&lt;&lt;"on-confirm"&gt;&gt;],
                    [&lt;&lt;"trust-user-id"&gt;&gt;,false],
                    [&lt;&lt;"uri"&gt;&gt;,&lt;&lt;"amqp://127.0.0.1:5672"&gt;&gt;]] in up-1
Key "uri" not found in up-1
 (//federation-upstream/up-1)
```

This issue is not present with a file produced by the HTTP API.

Discovered while working on https://github.com/rabbitmq/rabbitmq-server/issues/2384.

Apparently it was introduced by this change https://github.com/rabbitmq/rabbitmq-server/commit/460a432 that makes exported parameter values to be proplists and not maps.

michaelklishin:
otherwise, it would be serialized as a list of pairs, which is
not the format the import path (or an operator) expects.

Closes #435.

### DIFF
--- a/lib/rabbitmq/cli/ctl/commands/export_definitions_command.ex
+++ b/lib/rabbitmq/cli/ctl/commands/export_definitions_command.ex
@@ -129,7 +129,15 @@ defmodule RabbitMQ.CLI.Ctl.Commands.ExportDefinitionsCommand do
   # Implementation
   #
 
-  defp serialise(map, "json") do
+  defp serialise(raw_map, "json") do
+    # make sure all runtime parameter values are maps, otherwise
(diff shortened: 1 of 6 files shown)


michaelklishin:
Backported to `v3.8.x`.

PR #437, opened 2020-06-29
Make --help, help usage help messages consistent

michaelklishin:
* `help` printed an "Error:" at the top for no reason
 * `help [command]` with a non-existent command did not offer a suggestion
   like an attempt to invoke a non-existent command would
 * Exit codes were not consistent
 * `help --list-commands` had line break issues

With this change,

  * `help` is consistent with --help
  * `help [command]` is consistent with `[command] --help`
  * If a command is not found, either during the execution flow
    or `help [commnad]`, we consistently attempt a Jaro distance suggestion
  * Successful or effectively successful exits from the `help` command do not produce any error messages at the top


michaelklishin:
Backported to `v3.8.x`.

PR #440, opened 2020-06-30
if a node is the only cluster member, pass on the mirror sync critical check

michaelklishin:
as it would produce a confusing false positive.

Per discussion with @harshac, @ansd.

Closes rabbitmq/rabbitmq-cli#432


michaelklishin:
Backported to `v3.8.x`.

PR #446, opened 2020-07-04
Introduce 'rabbitmq-diagnostics erlang_cookie_sources'

michaelklishin:
to help troubleshoot authentication issues.

I could not quickly locate an Erlang function that would return the expected path in OTP so
we reinvent what [this BIF does](https://github.com/erlang/otp/blob/master/lib/erl_interface/src/connect/ei_connect.c#L2942) in C to cover the Windows-specific special handling of the `HOMEDRIVE` and `HOMEPATH` env variables.

Example output:

```
Cookie File

Effective user: user
Effective home directory: /Users/user
Cookie file path: /Users/user/.erlang.cookie
Cookie file exists? true
Cookie file type: regular
Cookie file access: read
Cookie file size: 20

Cookie CLI Switch

--erlang-cookie value set? false
--erlang-cookie value length: 0

Env variable  (Deprecated)

RABBITMQ_ERLANG_COOKIE value set? false
RABBITMQ_ERLANG_COOKIE value length: 0
```

Inspired by an idea from @gerhard.


michaelklishin:
Backported to `v3.8.x`.

PR #452, opened 2020-09-15
Quorum queue peek command

kjnilsson:
Requires: https://github.com/rabbitmq/rabbitmq-server/pull/2447

### DIFF
--- a/lib/rabbitmq/cli/queues/commands/peek_command.ex
+++ b/lib/rabbitmq/cli/queues/commands/peek_command.ex
@@ -10,8 +10,7 @@ defmodule RabbitMQ.CLI.Queues.Commands.PeekCommand do
   @behaviour RabbitMQ.CLI.CommandBehaviour
   def scopes(), do: [:queues]
 
-  def merge_defaults(args, opts), do: {args, Map.merge(%{vhost: "/"}, opts)}
-
+  use RabbitMQ.CLI.Core.MergesDefaultVirtualHost
   use RabbitMQ.CLI.Core.AcceptsTwoPositionalArguments
   use RabbitMQ.CLI.Core.AcceptsDefaultSwitchesAndTimeout
   use RabbitMQ.CLI.Core.RequiresRabbitAppRunning
--- a/lib/rabbitmq/cli/queues/commands/peek_command.ex
+++ b/lib/rabbitmq/cli/queues/commands/peek_command.ex
@@ -10,9 +10,30 @@ defmodule RabbitMQ.CLI.Queues.Commands.PeekCommand do
   @behaviour RabbitMQ.CLI.CommandBehaviour
   def scopes(), do: [:queues]
 
-  use RabbitMQ.CLI.Core.MergesDefaultVirtualHost
-  use RabbitMQ.CLI.Core.AcceptsTwoPositionalArguments
   use RabbitMQ.CLI.Core.AcceptsDefaultSwitchesAndTimeout
+
+  use RabbitMQ.CLI.Core.MergesDefaultVirtualHost
+
+  def validate(args, _) when length(args) < 2 do
+    {:validation_failure, :not_enough_args}
+  end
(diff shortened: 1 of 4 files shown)


lib/rabbitmq/cli/queues/commands/peek_command.ex
```diff
@@ -0,0 +1,53 @@
...
+##
+## Copyright (c) 2007-2020 VMware, Inc. or its affiliates.  All rights reserved.
+
+defmodule RabbitMQ.CLI.Queues.Commands.PeekCommand do
+  alias RabbitMQ.CLI.Core.DocGuide
+
+  @behaviour RabbitMQ.CLI.CommandBehaviour
+  def scopes(), do: [:diagnostics, :queues]
+
+  def merge_defaults(args, opts), do: {args, Map.merge(%{vhost: "/"}, opts)}
```
michaelklishin, line 13: This virtual host is an implicit default. Using `MergsNoDefaults` here would be functionally equivalent.

lib/rabbitmq/cli/queues/commands/peek_command.ex
```diff
@@ -0,0 +1,53 @@
...
+
+  use RabbitMQ.CLI.Core.AcceptsTwoPositionalArguments
+  use RabbitMQ.CLI.Core.AcceptsDefaultSwitchesAndTimeout
+  use RabbitMQ.CLI.Core.RequiresRabbitAppRunning
+
+  def run([name, pos] = _args, %{node: node_name, vhost: vhost}) do
+    {pos, _} = Integer.parse(pos)
```
michaelklishin, line 20: We should validate that the position is a positive integer.

lib/rabbitmq/cli/queues/commands/peek_command.ex
```diff
@@ -0,0 +1,53 @@
...
+        err
+    end
+  end
+
+  use RabbitMQ.CLI.DefaultOutput
+
+  def formatter(), do: RabbitMQ.CLI.Formatters.PrettyTable
+
+  def usage() do
+    "peek [--vhost <vhost>] <queue> <position>"
```
michaelklishin, line 37: Virtual host is documented in the general parameters, it is not specific to this command.

lib/rabbitmq/cli/queues/commands/peek_command.ex
```diff
@@ -0,0 +1,53 @@
...
+  @behaviour RabbitMQ.CLI.CommandBehaviour
+  def scopes(), do: [:diagnostics, :queues]
+
+  def merge_defaults(args, opts), do: {args, Map.merge(%{vhost: "/"}, opts)}
+
+  use RabbitMQ.CLI.Core.AcceptsTwoPositionalArguments
+  use RabbitMQ.CLI.Core.AcceptsDefaultSwitchesAndTimeout
+  use RabbitMQ.CLI.Core.RequiresRabbitAppRunning
+
+  def run([name, pos] = _args, %{node: node_name, vhost: vhost}) do
```
michaelklishin, line 19: We try to support JSON output where it's reasonably easy for newly added commands.

test/queues/peek_command_test.exs
```diff
@@ -0,0 +1,45 @@
...
+
+  test "validate: treats no arguments as a failure" do
+    assert @command.validate([], %{}) == {:validation_failure, :not_enough_args}
+  end
+
+  test "validate: treats a single positional argument as a failure" do
+    assert @command.validate(["quorum-queue-a"], %{}) == {:validation_failure, :not_enough_args}
+  end
+
+  test "validate: when two or more arguments are provided, returns a failure" do
```
michaelklishin, line 35: Two is a valid set of arguments. We should validate the position's data type and range, though.

michaelklishin (changes requested):

michaelklishin (approved):

michaelklishin:
Backported to `v3.8.x`.

PR #462, opened 2020-10-25
Provide an extra hint in 'rabbitmqctl add_user' output

michaelklishin:
to give the user an idea that she would have to
grant permissions to the newly added user.


lukebakken (commented):
:+1:

